PR #26, opened 2026-04-18
feat(opensearch-plugin): add @vendure-community/opensearch-plugin

LeftoversTodayAppAdmin:
# feat(opensearch-plugin): add `@vendure-community/opensearch-plugin`

## Summary

Adds a new community plugin, `@vendure-community/opensearch-plugin`, that powers
product search via [OpenSearch](https://opensearch.org/) (the Apache-2.0 community
fork of Elasticsearch). It is a near drop-in replacement for
`@vendure-community/elasticsearch-plugin` — GraphQL schema, option names, index
naming, and event handling are identical, so a storefront or admin UI built
against the ES plugin works unchanged.

The existing `@vendure-community/elasticsearch-plugin` is **not modified** by
this PR. Users pick whichever backend their stack targets.

## Motivation

OpenSearch is fully open source under Apache 2.0, while newer Elasticsearch
versions ship under the Elastic/SSPL licenses that restrict managed-service
offerings. More importantly for our use case, OpenSearch offers first-class
**cross-cluster replication** (CCR) for basic deployments, which we rely on in
the LeftoversToday stack to replicate product data across geographies without
an enterprise Elasticsearch subscription.

Giving the Vendure community a supported OpenSearch plugin lets teams on the
OS fork keep using the same battle-tested search architecture from the
Elasticsearch plugin — without maintaining a local fork.

## What's included

A brand-new package at `packages/opensearch-plugin/` with the full feature set
of the ES plugin:

- `OpenSearchPlugin` Vendure plugin class with `.init(options)` and
  `.ui` helpers
- `OpenSearchService` / `OpenSearchIndexerController` mirroring the ES
  indexer — same queue name (`'update-search-index'`), same event
  subscriptions, same reindex semantics
- Shop & Admin GraphQL resolvers (`Shop/Admin/EntityOpenSearchResolver`)
  exposing the same `search` / `searchCollections` fields the ES plugin
  exposes today
- Custom-mappings and custom-script-fields resolvers
- Health indicator (`opensearch.health.ts`) — the `@deprecated` annotation
  from the ES plugin is preserved, matching upstream guidance
- Buffered-updates support, channel-aware indexing, language-aware search,
  `groupByProduct` / `groupBySKU`, price-range aggregations, facet-value
  aggregations — all ported verbatim
- Admin UI shell (same extension surface as the ES plugin)

## Test coverage (parity with ES plugin)

Full 1:1 mapping against the ES plugin's test suite — verified with
`diff <(find packages/elasticsearch-plugin/{src,e2e} -type f | sort)
       <(find packages/opensearch-plugin/{src,e2e} -type f | sort)` (only
filename renames differ).

| Suite                                   | Count |
| --------------------------------------- | ----- |
| Unit — `src/build-search-body.spec.ts`  | 33    |
| E2E — `e2e/opensearch-plugin.e2e-spec.ts` | 91    |
| E2E — `e2e/opensearch-plugin-uuid.e2e-spec.ts` | 5 |
| **Total**                               | **129** |

All pass locally and on a clean Ubuntu box against a fresh
`opensearchproject/opensearch:3.0.0` container:

## Server compatibility

- **Targets:** OpenSearch **2.19+** and **3.x** servers
- **Client:** `@opensearch-project/opensearch@^3.5.1`
- Tested against `opensearchproject/opensearch:3.0.0`

## API differences handled in the port

The `@opensearch-project/opensearch` client is close to, but not identical to,
`@elastic/elasticsearch`. The following differences are handled internally
(users do not see them):

1. **Response envelope is always `{ body, statusCode, headers }`.** The
   `{ meta: true }` option from the ES 8+ client is removed everywhere; all
   call sites read `result.body`.
2. **Payloads live under `body:`.** `mappings`, `settings`, `query`, `script`,
   `actions`, etc. are all wrapped inside a `body` parameter on the client
   call.
3. **`client.search()` takes no generic.** Unlike the ES client, a generic
   type arg on `search<T>()` fails with `TS2558`. The port casts the body
   through `unknown` to the expected `SearchResponseBody<VariantIndexItem>`.
4. **`client.updateByQuery()` response is a union type.** The OS client's
   `UpdateByQuery_ResponseBody` includes the async-task variant
   `{ task?: string }`, so `.failures` is not statically available. The port
   extracts an `any`-typed local (`body1`/`body2`) to access `.failures` on
   the synchronous path we actually take.
5. **Error shape.** Error handling uses `e.meta?.body?.error?.type` with
   optional chaining to tolerate shape drift between client minor versions.

### `docker-compose.yml`

Adds an `opensearch` service alongside the existing `elasticsearch` service so
both can run side-by-side during development:

```yaml
opensearch:
    image: opensearchproject/opensearch:3.0.0
    environment:
        - cluster.name=opensearch-docker-cluster
        - node.name=os
        - discovery.type=single-node
        - bootstrap.memory_lock=true
        - 'OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m'
        - 'DISABLE_SECURITY_PLUGIN=true'
        - 'DISABLE_INSTALL_DEMO_CONFIG=true'
    ulimits:
        memlock: { soft: -1, hard: -1 }
    volumes:
        - osdata:/usr/share/opensearch/data
    ports:
        # Host 9201 -> container 9200 so it can run alongside elasticsearch.
        - 9201:9200
```

## Usage

```jsx
import { OpenSearchPlugin } from '@vendure-community/opensearch-plugin';

export const config: VendureConfig = {
    plugins: [
        OpenSearchPlugin.init({
            host: 'http://localhost',
            port: 9200,
            // every other option identical to ElasticsearchPlugin
            indexPrefix: 'vendure-',
            batchSize: 2000,
            // ...
        }),
    ],
};
```

To try it against the dev-server:

```
docker compose up -d opensearch       # host 9201 -> container 9200
uncomment OpenSearchPlugin import + init in dev-server/dev-config.ts
npm run populate
npm run dev
```

## Breaking changes
None. This is a brand-new package at packages/opensearch-plugin/. Nothing in
elasticsearch-plugin, dev-server, or the root tooling is altered in a way
that affects existing consumers.


michaelbromley:
Hey there,

So I don't think this is the way to go - there's 98% code duplication here with the ES plugin, which is not good for maintainability at all.

A better approach would be to modify the ES plugin to make the actual search client part pluggable. Suggestion:

  - Extract a thin SearchClientAdapter interface with methods like search, indices.exists, updateByQuery, existsAlias, getMapping, etc. — all returning
   a normalised { body } envelope.
  - Ship ElasticsearchAdapter and OpenSearchAdapter implementations (maybe 150–250 lines each) in the same package.
  - Let users pick via init({ client: 'elasticsearch' | 'opensearch', … }) or by passing an adapter instance.
  - All the service, indexer, resolvers, GraphQL schema, e2e fixtures, options stay shared.

I'll close this, but I'm certainly open to a new PR that does something like what I suggest above.

LeftoversTodayAppAdmin:
Hi @michaelbromley!

Fair point, and thanks for the direction, the adapter pattern can also work and I originally stayed away from it to avoid breaking changes to the existing plugin and not get stuck with the lowest common denominator of features between them. I'll redo it the way you have suggested. 

Before I start, a few clarifications so I come back with something mergeable on the first try:

- Package identity: keep @vendure-community/elasticsearch-plugin and the ElasticsearchPlugin class name, with OpenSearch as an opt-in via a new option? Or would you prefer a neutral rename (e.g. search-plugin) and deprecate the old name? Happy either way, just want to get it right once.
- Selection API: are you thinking init({ client: 'elasticsearch' | 'opensearch', ... }) as a string discriminator, or strictly an adapter instance with two exported factories (createElasticsearchAdapter, createOpenSearchAdapter)? I'd slightly lean toward supporting both, string for the 95% case, adapter instance as an escape hatch for tests and custom subclasses, but will follow your preference.
- Dependencies: both clients as peerDependencies so users only install the one they use, with a friendly error if the selected adapter's client package is missing?
- Divergence escape hatches: to keep the adapter from becoming a lowest-common-denominator trap as OS and ES drift apart (OS k-NN/CCR, ES ESQL, etc.), I'd like to build in two things from day one: (a) typed per-backend option blocks on init(), elasticsearch: {...} and opensearch: {...}, where each block is only read when that client is selected, and (b) a getRawClient() method on the adapter interface so advanced users can reach the native client for anything we don't model. OK with that direction?
- Tests: plan is to parameterize the existing e2e suite over both adapters (vitest describe.each) running against the ES and OS services side-by-side in CI, separate index prefixes so they can't collide.
- If the answers are "keep the package, string discriminator with adapter escape hatch, peer deps, per-backend option blocks and getRawClient, parameterised e2e", I'll open the new PR. Happy to do a tiny design-only PR with just the SearchClientAdapter interface first if you'd rather lock that down before I touch the service and indexer code.

michaelbromley:
1. keep the package name the same, but we'll just make clear in the docs that it supports both.
2. adapter sounds like the way to go, let's see what the implementation looks like and I'll be able to judge better.
3. yes
4. could the adapter-specific config not be passed to the adapter factory instead of the top level init options? Get raw client sounds like a good idea
5. ok

LeftoversTodayAppAdmin:
@michaelbromley  Perfect, that's all I needed. Moving adapter-specific config into the factory is actually cleaner, agreed. I'll open a draft PR with the SearchClientAdapter interface, both adapter factories, getRawClient, peer-deps with a friendly "install X" error, and the parameterized e2e suite running both backends. Will ping you here once it's up.

PR #27, opened 2026-04-23
feat(elasticsearch-plugin)!: add pluggable SearchClientAdapter for Elasticsearch or OpenSearch - Breaking Change

LeftoversTodayAppAdmin:
## Summary

Replaces the hard-coded `@elastic/elasticsearch` dependency in `@vendure-community/elasticsearch-plugin` with a pluggable `SearchClientAdapter` interface, so the same plugin can be backed by either Elasticsearch or OpenSearch by passing an adapter factory at `init()` time.

This is an alternative to shipping a separate `@vendure-community/opensearch-plugin` (as explored in #26): a single plugin, two first-party adapters, and both clients declared as _optional_ peer dependencies so consumers only install the one they use.

## What changed

**Public API**

- New `SearchClientAdapter` interface (exported from the package entry).
- Two first-party adapter factories:
  - `createElasticsearchAdapter({ host, port, clientOptions })` — backed by `@elastic/elasticsearch`.
  - `createOpenSearchAdapter({ host, port, clientOptions })` — backed by `@opensearch-project/opensearch`.
- `ElasticsearchPlugin.init({ adapter, ...pluginOptions })` — `adapter` is a **factory** `() => SearchClientAdapter`. The plugin invokes it once per internal NestJS provider (read-side service + write-side indexer controller), so each provider owns an independent client & connection pool. A single shared instance would be torn down twice during `onModuleDestroy` and take both providers offline.
- Each adapter exposes `adapter.getRawClient()` as an escape hatch for queries that aren't on the `SearchClientAdapter` surface.
- Consumers can also implement `SearchClientAdapter` themselves (e.g. for a managed/hosted service with a bespoke SDK, or a test double) and pass a factory that returns fresh instances.

**Dependencies**

- `@elastic/elasticsearch` and `@opensearch-project/opensearch` are now both declared as `peerDependencies` with `peerDependenciesMeta.optional = true`.
- Minimum `@elastic/elasticsearch` tightened to `^9.1.0` to align with Vendure `3.6.0`'s [minimum Elasticsearch requirement](https://github.com/vendurehq/vendure/releases/tag/v3.6.0).
- OpenSearch client/server `3.x` line supported.

**Monorepo**

- `packages/opensearch-plugin/` deleted — the adapter pattern obsoletes the separate package.
- CI (`.github/workflows/build-and-test.yml`) updated: the `elasticsearch-plugin` e2e job is now a matrix that runs the same spec file against a dedicated Elasticsearch service and a dedicated OpenSearch service.

**Docs**

- Plugin `README.md` rewritten for the adapter API, including a "Migrating from v1.x" section with a worked example.
- `CHANGELOG.md` gets a `2.0.0` entry describing the new API and the breaking change.

## Breaking changes

`ElasticsearchPlugin.init({ host, port, clientOptions })` is replaced with `ElasticsearchPlugin.init({ adapter: () => SearchClientAdapter })`. Existing configuration migrates like this:

**Before (v1.x):**

```ts
ElasticsearchPlugin.init({
  host: 'http://localhost',
  port: 9200,
  clientOptions: { auth: { username: 'elastic', password: 'changeme' } },
  indexPrefix: 'vendure-',
});
```

**After (v2.x):**

```ts
import { ElasticsearchPlugin, createElasticsearchAdapter } from '@vendure-community/elasticsearch-plugin';

ElasticsearchPlugin.init({
  adapter: () =>
    createElasticsearchAdapter({
      host: 'http://localhost',
      port: 9200,
      clientOptions: { auth: { username: 'elastic', password: 'changeme' } },
    }),
  indexPrefix: 'vendure-',
});
```

Note the arrow wrapper: `adapter` takes a factory that the plugin calls once per internal provider. The factory shape is required — sharing a single adapter instance across providers would cause one provider's `onModuleDestroy` to drain the shared client mid-flight in the other.

All other top-level options on `ElasticsearchOptions` (`indexPrefix`, `connectionAttempts`, `customProductMappings`, `customFields`, etc.) are unchanged.

## Test plan

- [x] Unit suite green (`bun run test` in `packages/elasticsearch-plugin`).
- [x] `tsc -p tsconfig.build.json` clean.
- [x] `elasticsearch-plugin.e2e-spec.ts` + `elasticsearch-plugin-uuid.e2e-spec.ts` — **96/96 passing** against Elasticsearch 9.1 on port 9200 (`SEARCH_BACKEND=elasticsearch`).
- [x] `elasticsearch-plugin.e2e-spec.ts` + `elasticsearch-plugin-uuid.e2e-spec.ts` — **96/96 passing** against OpenSearch 3.0 on port 9201 (`SEARCH_BACKEND=opensearch`).
- [x] Full-repo `bun run e2e` from the repo root (matches CI's matrix, runs all 4 packages concurrently via `lerna run e2e`) green on both backends.
- [x] Clean teardown on both backends (no hanging `afterAll`).

### Test-infra fixes rolled into this PR

- Extracted `buildAdapterForBackend` into its own e2e module so the two spec files don't transitively import each other's top-level `ElasticsearchPlugin.init()` and clobber the plugin's static options mid-worker.
- Made adapter `close()` idempotent (memoised promise) — a cheap belt-and-braces guard even though the factory-per-provider architecture means each provider now owns a separate client and the double-close contention it originally mitigated is gone.
- Bumped the `afterAll` hook timeout to `TEST_SETUP_TIMEOUT_MS` for teardown headroom under OpenSearch.


LeftoversTodayAppAdmin:
Hi @michaelbromley! The new PR is ready for your review!

michaelbromley:
Hey @LeftoversTodayAppAdmin - thanks for this! I’ve just been doing some housekeeping on this repo and part of it was updating to Bun as a package manager. Can you take a look at the conflicts and resolve and then ping me again? Cheers.

LeftoversTodayAppAdmin:
@michaelbromley rebased onto upstream/main (over the bun migration). Workflow is ported to `bun install --frozen-lockfile` / `bun run build|test|lint|e2e` matching the new convention you set; matrix kept on `[20.x, 22.x] × [elasticsearch, opensearch]`. `bun.lock` regenerated and folded into the adapter feat commit alongside the new optional peer-dep entries.

CI green across all 4 matrix jobs (96/96 e2e on each backend). Verified locally on the same versions (`bun 1.3.10`, Node `20.19.0`) before pushing.

Ready for another look when you have a moment.

michaelbromley (commented):
Is the two-instance design primarily about connection-pool isolation between read/write paths? 

The changelog mentions double-teardown, but the memoised closePromise already handles that safely. If pool isolation is the real intent, mind updating the rationale in the changelog/README so future maintainers understand it?

LeftoversTodayAppAdmin:
@michaelbromley Before this fix, the tests were randomly failing in different orders so I was trying to figure out why that was happening and didn't fully update the documentation for the fix, let me do that now.

Quick walkthrough of why we hand each provider its own client:

**1. Connection-pool isolation between read and write paths.** This is the primary reason. The plugin has two long-lived NestJS providers that touch the search backend:

- `ElasticsearchService` - read-side, serves the user-facing `search` query and the `searchProductsAdmin` resolvers. Latency-sensitive.
- `ElasticsearchIndexerController` - write-side, runs bulk reindex / `updateByQuery` / `deleteByQuery`, often via the job queue. Throughput-heavy and bursty.

If both share one client, a long reindex (or even a large `bulk` payload) can occupy every connection in the underlying pool and stall live search responses behind it. Giving each provider its own client + pool means bulk indexing traffic never competes with user query traffic for sockets.

**2. Shutdown ordering.** Both providers implement `onModuleDestroy` and each closes its own client. With a shared instance, whichever provider Nest happens to destroy first would tear the underlying client out from under the other one mid-flight (in-progress reindex jobs, or a still-streaming search response). With separate instances each provider drains independently.

**About `closePromise`:** you're correct that it makes a second `close()` a safe no-op, but that's a defensive guard for the case where someone misconfigures the factory to return the same instance (e.g. `const a = createElasticsearchAdapter(...); ElasticsearchPlugin.init({ adapter: () => a })`). It's not the architectural reason for the factory; the architectural reason is the two points above. The new wording in the changelog calls that out explicitly so future readers don't conflate the two. 

CI should re-run on the new commit shortly. Let me know if the rewrite reads clearly or if you'd like it phrased differently. If you would rather I remove the second close() in closePromise, happy to do that as well.

michaelbromley (approved):

michaelbromley:
Thank you for your work on this!

PR #33, opened 2026-05-06
fix(mollie-payments): wrap addPayment in transaction

casperiv0:
This PR includes:

With the recent release of Vendure, it requires the `transitionOrderToState` and `addPaymentToOrder` fns to be wrapped in a Transaction. 
That wasn't handled inside the `addPayment` helper function in the Mollie service.

cc @martijnvdbrug would be nice to have your input on this too, please 🙏

Related Vendure Core PR: https://github.com/vendurehq/vendure/pull/4689

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/packages/mollie-plugin/src/mollie.service.ts
+++ b/packages/mollie-plugin/src/mollie.service.ts
@@ -377,6 +377,8 @@ export class MollieService {
         paymentMethodCode: string,
         status: 'Authorized' | 'Settled',
     ): Promise<Order> {
+        // `addPaymentToOrder` requires a db transaction. Free orders use a new
+        // admin `RequestContext` in `createPaymentIntent`, which does not keep the transaction.
         return this.connection.withTransaction(_ctx, async ctx => {
             if (order.state !== 'ArrangingPayment' && order.state !== 'ArrangingAdditionalPayment') {
                 const transitionToStateResult = await this.orderService.transitionToState(
--- a/packages/mollie-plugin/src/mollie.service.ts
+++ b/packages/mollie-plugin/src/mollie.service.ts
@@ -370,17 +370,17 @@ export class MollieService {
      * Add payment to order. Can be settled or authorized depending on the payment method.
      */
     async addPayment(
-        ctx: RequestContext,
+        _ctx: RequestContext,
         order: Order,
         amount: number,
         mollieMetadata: Omit<MolliePaymentMetadata, 'status' | 'amount'>,
         paymentMethodCode: string,
         status: 'Authorized' | 'Settled',
     ): Promise<Order> {
-        return this.connection.withTransaction(ctx, async txCtx => {
+        return this.connection.withTransaction(_ctx, async ctx => {
             if (order.state !== 'ArrangingPayment' && order.state !== 'ArrangingAdditionalPayment') {
                 const transitionToStateResult = await this.orderService.transitionToState(
-                    txCtx,
+                    ctx,
                     order.id,
                     'ArrangingPayment',
                 );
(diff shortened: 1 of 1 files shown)


packages/mollie-plugin/src/mollie.service.ts
```diff
@@ -375,37 +377,39 @@ export class MollieService {
...
-            const transitionToStateResult = await this.orderService.transitionToState(
-                ctx,
-                order.id,
-                'ArrangingPayment',
-            );
-            if (transitionToStateResult instanceof OrderStateTransitionError) {
-                throw Error(
-                    `Error transitioning order ${order.code} from ${transitionToStateResult.fromState} ` +
-                        `to ${transitionToStateResult.toState}: ${transitionToStateResult.message}`,
+        return this.connection.withTransaction(ctx, async txCtx => {
```
martijnvdbrug: Maybe we can rename the original ctx `_ctx` or something, to prevent accidental usage of the original ctx. `_ctx` would indicate this isn't a normal request context.

martijnvdbrug (changes requested):
Minor optional change requested, but approved if you choose to ignore it :-) 

Good catch. I have yet to test it out though...

packages/mollie-plugin/src/mollie.service.ts
casperiv0: Thanks for the suggestion!

casperiv0:
Thanks Martijn! Tested in my Vendure project w/ our test suite and are passing now after this change. Before they were failing, mentioning that the fn wasn't wrapped in a transaction

martijnvdbrug (approved):

grolmus (approved):
Reviewed locally + ran build & lint, both green.

The wrap pattern is correct and matches the canonical Vendure idiom mirroring core #4689. `withTransaction` correctly joins an existing tx on `_ctx` (webhook path via `@Transaction()` on `MollieController`) and opens a new one when `_ctx` has none (free-order path in `createPaymentIntent` where a fresh admin `RequestContext` is built). Applying the wrap at `addPayment` rather than higher up in `handleMolliePaymentStatus` is the right call — it also protects the free-order path and any future direct callers.

A few non-blocking suggestions for a follow-up:

1. **Param naming**: `_ctx` is a touch confusing — the leading underscore conventionally signals an unused parameter, but here it's passed to `withTransaction`. Something like `outerCtx` would read better. Minor.

2. **Comment clarity** at the top of `addPayment`: the current note mentions the free-order admin ctx specifically, but `addPayment` is a public method and any future caller without an outer tx would hit the same issue. Could be reworded to lead with the general invariant — "`transitionToState` and `addPaymentToOrder` must run in a single tx so save/hooks/save commit or roll back together; `withTransaction` joins an existing tx on the passed ctx, or opens a new one (e.g. for the fresh admin ctx in `createPaymentIntent` free-order branch)."

3. **E2E regression**: core #4689 added `order-process-rollback.e2e-spec.ts` as a regression guard. A sibling spec in `packages/mollie-plugin/e2e/` that registers a throwing hook and asserts no half-transitioned state after `addPayment` failure would protect against future refactors. Worth tracking as a follow-up issue.

LGTM, approving.

PR #36, opened 2026-06-08
feat(stripe-plugin)!: modernize for Stripe SDK 22 / dahlia API (OSS-555 items 1-3, 6)

michaelbromley:
Closes [OSS-555](https://linear.app/vendure/issue/OSS-555) — scoped to items **1, 2, 3, 6** (pure modernization). Items **4, 5, 7** (async/processing webhook, dashboard refund + dispute sync, SetupIntents) are deferred to follow-up issues.

## What's in this PR

| Commit | OSS-555 item | Summary |
|--------|--------------|---------|
| `bdc9035` | 1 | Stripe SDK `13.x` → `^22.0.0` peer + dev. Local `src/stripe-types.ts` shim works around [stripe-node#2683](https://github.com/stripe/stripe-node/issues/2683) (CJS NodeNext default import doesn't expose namespace types). |
| `e58cc1b` | 2 | `apiVersion` defaults to the SDK pinned version (`2026-05-27.dahlia`). New plugin option lets callers pin a specific version or set `null` to fall back to the Stripe account default (pre-2.0.0 behaviour). |
| `392a493` | 3 | `nock` 13 → 14, `rimraf` 5 → 6. `typescript` left at 5.8.2 (monorepo-pinned). |
| `6115123` | 6 | `createRefund` forwards Vendure's refund `reason` to Stripe (mapped to the enum where possible, otherwise stored on `metadata.vendureRefundReason`) and adds an idempotency key. |
| `e0a6606` | — | Tests adapted: switch `VendureStripeClient` to `Stripe.createFetchHttpClient()`. nock 14 (via `@mswjs/interceptors`) deadlocks Stripe's `NodeHttpClient` — see test plan below. |
| `a1381fc` | — | Hand-written 2.0.0 CHANGELOG section + README pointer. |

`feat!` on the apiVersion commit triggers the 2.0.0 major bump via the release flow.

## Verified

- `bun run build` clean across all 8 packages.
- `bun run lint` clean (0 errors, 12 pre-existing warnings).
- `bun run e2e` in `stripe-plugin`: 18/18 pass.

## Not test-covered — needs reviewer eyeball

These three behavioural changes are not covered by the existing e2e suite (the suite mocks Stripe entirely via nock):

1. **HTTP transport switched from `node:http(s)` to `fetch`.** Production users who inject a custom `http.Agent` (proxy, keep-alive tuning) will silently lose it. An `httpClient` escape hatch is planned as a follow-up commit on this PR.
2. **Default API version pinned to `dahlia`.** Stripe API responses on this account will now be shaped per the 2026-05-27 API. Per the Stripe changelog the core `PaymentIntent` / `Refund` / `Charge` / `Customer` shapes are stable between `2023-08-16` and `dahlia`, but this is unverified against a live account.
3. **Refund metadata fallback.** A non-Stripe-enum reason string ends up on `metadata.vendureRefundReason`, not the structured `reason` field.

---

### 1. Risk × evidence matrix

| Risk | What could break | Doc evidence | Verification plan |
|------|------------------|--------------|-------------------|
| Default API version moves `2023-08-16` → `2026-05-27.dahlia` (~3 years of evolution) | Field shapes on objects we touch (\`PaymentIntent\`, \`Refund\`, \`Charge\`, webhook \`Event\`) could shift. | Stripe's API changelog confirms that across \`2023-10-16\`, \`2024-04-10\`, \`2024-06-20\`, acacia (\`2024-09-30\`), basil (\`2025-03-31\`), clover and dahlia, **no breaking changes were made to the PaymentIntent / Refund / Charge / Customer shapes used by this plugin** — breaking changes in that window are scoped to Connect, Issuing, Terminal, Billing/Subscriptions, and Stripe.js method renames ([changelog/dahlia.md](https://docs.stripe.com/changelog/dahlia.md), [changelog/basil.md](https://docs.stripe.com/changelog/basil.md), [changelog/2024-06-20](https://docs.stripe.com/changelog/2024-06-20)). Only forward-compatible additions to enums like \`PaymentMethod.type\` (\`bizum\`, \`scalapay\`, \`sunbit\`, \`pix\`, etc., per stripe-node v22.1/22.2 changelog) — we don't switch on \`type\`, so this is moot for us. | (a) Walk the diff of \`node_modules/stripe/cjs/resources/{PaymentIntents,Refunds,Charges,Customers,Events}.d.ts\` against the same file in stripe@13.11.0 and confirm only additive changes touch the fields we read (\`id\`, \`client_secret\`, \`status\`, \`amount_received\`, \`last_payment_error.message\`, \`metadata\`, \`failure_reason\`). (b) During the manual walkthrough below, log the raw webhook payload and the \`paymentIntents.create\` response to confirm shape. |
| Webhook signature verification under new SDK | \`constructEvent\` could change semantics. | [webhooks/signatures](https://docs.stripe.com/webhooks/signatures): the v1 HMAC-SHA256 + \`Stripe-Signature\` scheme is **API-version-agnostic** and unchanged. stripe-node 22.0.1 only added \`string[]\` support to the signature param. Existing 18 e2e tests already exercise the signed and unsigned paths. | Already covered by \`e2e/stripe-payment.e2e-spec.ts\` ("Should not crash on signature validation failure", "Should validate the webhook's signature properly"). Verify in manual run with a *real* Stripe-signed webhook from \`stripe listen\`. |
| Webhook events for an older account API version still arrive in old shape | Even though we pin \`dahlia\` for outbound, inbound webhook events use the API version configured on the webhook endpoint at the Stripe dashboard (or the account default). A merchant on \`2023-08-16\` sends \`2023-08-16\`-shaped events. | [Stripe upgrades doc](https://docs.stripe.com/upgrades): events use the API version active at event creation; this is not affected by client \`apiVersion\`. The plugin only reads stable fields, so OK. | Manual walkthrough done twice: once with the Stripe account's default API version, once with the webhook endpoint pinned to a newer version (configurable in dashboard). |
| SDK v13 → v22 runtime contract | Removed callback support, ES6 class constructor, per-request host override removed, \`Stripe.errors.StripeError\` is now value-only. | stripe-node [v22 migration guide](https://github.com/stripe/stripe-node/wiki/Migration-guide-for-v22) + local \`node_modules/stripe/CHANGELOG.md\` lines for 22.0.0. We don't use callbacks, host overrides, or per-request API keys; we already use \`new Stripe(...)\` and \`instanceof Stripe.errors.StripeError\` (value access, unaffected by the type-only change). | Build (passes) + e2e (18/18 pass). Add a unit-ish test that asserts \`Stripe.errors.StripeError\` is still a constructable class so a future SDK drift fails loudly. |
| NodeHttpClient → FetchHttpClient | Custom \`http.Agent\` settings (proxy, keep-alive tuning) injected via deployment can't be applied to FetchHttpClient. Retry/timeout behaviour and error shapes may differ. | stripe-node ships both clients (\`createFetchHttpClient\` is the documented Worker/Edge path, and the SDK supports it equivalently in Node since v18). The deadlock with nock@14 is documented in the @mswjs/interceptors socket lifecycle (\`mockConnect()\` runs inside \`respondWith()\` — \`connect\`/\`secureConnect\` only fire post-match, see \`node_modules/@mswjs/interceptors/lib/node/ClientRequest-*.cjs\`). | (a) Add a \`httpClient\` option to \`StripePluginOptions\` so users who *do* need an \`http.Agent\` can pass their own. (b) Document the migration in CHANGELOG (done). (c) Manual smoke test confirms a real card payment + a real refund both work end-to-end. (d) Add an automated test that asserts \`FetchHttpClient\` is wired by inspecting \`(stripe as any)._api.httpClient.getClientName() === 'fetch'\`. |
| Refund \`reason\` mapping | We map free-form Vendure refund reason strings onto Stripe's three-value enum \`duplicate \| fraudulent \| requested_by_customer\`; everything else goes to \`metadata.vendureRefundReason\`. | Stripe Refunds API reference: \`reason\` is a strict enum at the API level. [api/idempotent_requests](https://docs.stripe.com/api/idempotent_requests) confirms parameter mismatch on a replayed key returns an error — so the mapping has to be deterministic for the same input. Our mapping is. | Add unit-level test against \`StripeService.createRefund\` (mocked Stripe) for each of: enum-matching reason, non-matching reason, no reason. Also assert the request body forwarded to nock. |
| Refund idempotency key format | We use \`refund_{payment_intent_id}_{amount}\` — deterministic. Stripe recommends UUIDs but accepts any string ≤255 chars; parameter mismatch on replay errors. | [api/idempotent_requests](https://docs.stripe.com/api/idempotent_requests): "the idempotency layer compares incoming parameters to those of the original request and errors if they're not the same"; keys expire after ~24h. Our key intentionally encodes the unique parameters (PI + amount), so two identical retries replay safely; two *different* refunds for the same PI+amount within 24h would collide, but that's vanishingly rare for the same payment. | Add a unit-level test that fires \`createRefund\` twice with the same args and asserts both calls send the same \`Idempotency-Key\`. Document the 24h+identical-params rule in CHANGELOG. |
| \`stripe-types.ts\` shim fragility | We extract types via \`Parameters<typeof stripe.refunds.create>[0]\` etc. If stripe-node restructures method overloads, our types shift silently. | The shim documents that it's a workaround for [stripe-node#2683](https://github.com/stripe/stripe-node/issues/2683) / [PR #2725](https://github.com/stripe/stripe-node/pull/2725); once that lands and we pin the version we delete the shim. | (a) Add a CI tripwire — a \`.ts\` file with \`expectType<>\` assertions on each shimmed type so a stripe SDK upgrade fails the build instead of silently changing meaning. (b) Watch the PR for merge; open a follow-up Linear issue to delete the shim when ready. |
| nock@14 / @mswjs/interceptors socket semantics | The mock's \`connect\`/\`secureConnect\` events only fire after \`respondWith()\` runs, deadlocking any client that gates body writes on \`secureConnect\`. Stripe's \`NodeHttpClient\` does that. | Source-confirmed: \`node_modules/@mswjs/interceptors/lib/node/ClientRequest-*.cjs::mockConnect()\` is only called from \`respondWith()\`. nock@14 [release notes](https://github.com/nock/nock/releases/tag/v14.0.0) only call out "drop Node <18" and "support for native fetch" as breaking — the socket lifecycle change isn't called out. | FetchHttpClient sidesteps it. Test suite proves it (18/18). |

### 2. Pre-merge actions (in order)

- [ ] **Fix \`e2e/stripe-dev-server.ts\`** — currently broken pre-existing TS errors (\`@vendure/admin-ui-plugin\` import, stale codegen exports). Without this we can't actually run the dev server, which is the *only* way to exercise real Stripe + real webhook.
- [ ] **Manual walkthrough** against a Stripe test account:
  - \`stripe listen --forward-to localhost:3000/payments/stripe\` (webhook signing secret goes into the plugin config).
  - **Happy path** — create order, request payment intent, pay with \`4242 4242 4242 4242\`, confirm order transitions to \`PaymentSettled\`, payment record's \`transactionId\` matches the PI id.
  - **Decline path** — pay with \`4000 0000 0000 0002\`. Confirm \`payment_intent.payment_failed\` is logged but order doesn't transition (matches today's behaviour; item 4 of OSS-555 fixes the surrounding gap separately).
  - **Refund path** — admin UI refunds the settled payment with a reason (test both a \`requested_by_customer\` string and a custom string). Confirm Stripe dashboard shows the refund with the right \`reason\` field and \`vendureRefundReason\` metadata respectively.
  - **Idempotency** — manually retry the refund call and confirm Stripe returns the same refund (idempotent replay), not a duplicate.
  - **apiVersion override** — restart with \`apiVersion: null\` set. Confirm requests go out without a \`Stripe-Version\` header (visible in \`stripe listen\` output). Then with \`apiVersion: '2024-06-20' as any\` — confirm header reflects it.
  - **Webhook payload shape** — log one full webhook payload in each run; eyeball that the fields the controller reads (\`id\`, \`metadata.{channelToken,orderId,orderCode,languageCode}\`, \`amount_received\`, \`status\`, \`last_payment_error?.message\`) are present and shaped as expected.
- [ ] **Add three targeted automated tests**:
  - \`createRefund\` forwards \`reason: 'requested_by_customer'\` when the input matches a Stripe enum value, and \`metadata.vendureRefundReason\` when it doesn't.
  - \`createRefund\` calls Stripe with the deterministic idempotency key (assert the \`Idempotency-Key\` header captured by nock).
  - \`apiVersion: null\` results in no \`Stripe-Version\` header sent; \`apiVersion: 'X'\` results in \`Stripe-Version: X\`.
- [ ] **Type-shim tripwire** — \`src/__type-checks.ts\` with \`expectAssignable<>\` asserts pinning a representative field on each shimmed type. Future SDK drift breaks the build instead of silently shifting meaning.
- [ ] **\`httpClient\` plugin option escape hatch** for production users who need an \`http.Agent\`. Default stays \`createFetchHttpClient()\`; if set, pass through.
- [ ] **Diff stripe@13.11.0 vs stripe@22.2.0 d.ts shapes** for: \`Charge\`, \`Refund\`, \`PaymentIntent\`, \`PaymentIntentCreateParams\`, \`Refund.status\`, \`Refund.failure_reason\`. Confirm fields the plugin reads are still defined the same way.

### 3. Post-merge

- Open follow-up Linear issues for: Items 4–7 of OSS-555; delete the type shim once [stripe-node#2725](https://github.com/stripe/stripe-node/pull/2725) ships; consider re-adding NodeHttpClient as the default once nock fixes the socket lifecycle.

### 4. Remaining assumptions after this plan

- Behaviour against payment methods we don't have test card numbers for in the walkthrough (BNPL, SEPA, bank transfer). Items 4–7 of OSS-555 are scoped to add coverage there.
- \`FetchHttpClient\` retry / timeout / connection-pool behaviour at high concurrency. We're trusting Stripe's own implementation; this is the same SDK code Stripe ships for Workers and Edge, so it has production miles, just not under this exact codepath.
- Subtle webhook payload differences between API versions that don't break our code today but might if Stripe later removes a field still hidden in test fixtures (the mocked webhook payloads in our tests are forged, not captured from real Stripe).

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/36"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783519280&installation_id=128408429&pr_number=36&repository=vendurehq%2Fcommunity-plugins&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fcommunity-plugins%2Fpull%2F36&signature=277b93e9ff528f11307a383d7412840b41b0b10259288fc6a8358c189e005eaf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -199,8 +199,8 @@
         "@vendure/common": "3.6.0-minor-202603280303",
         "@vendure/core": "3.6.0-minor-202603280303",
         "@vendure/testing": "3.6.0-minor-202603280303",
-        "nock": "^13.1.4",
-        "rimraf": "^5.0.5",
+        "nock": "^14.0.0",
+        "rimraf": "^6.0.0",
         "stripe": "^22.0.0",
         "typescript": "5.8.2",
       },
@@ -528,6 +528,8 @@
 
     "@mollie/api-client": ["@mollie/api-client@4.5.0", "", { "dependencies": { "@types/node-fetch": "^2.6.13", "node-fetch": "^2.7.0", "ruply": "^1.0.1" } }, "sha512-Aj3Erv3EqY74fFK8gYegjzXZxPNq4T7cbsoPueRx5aF6o0kL8kR9eeaLZPaFkv2zHA4XQW6ky8yG7CMyu4oqnw=="],
 
+    "@mswjs/interceptors": ["@mswjs/interceptors@0.41.9", "", { "dependencies": { "@open-draft/deferred-promise": "^2.2.0", "@open-draft/logger": "^0.3.0", "@open-draft/until": "^2.0.0", "is-node-process": "^1.2.0", "outvariant": "^1.4.3", "strict-event-emitter": "^0.5.1" } }, "sha512-VVPPgHyQ6ShqnrmDWuxjmUIsO9gWyOZFmuOfLd9LfBGQJwZfy0gvv9pbHSJuoFNIYC7ZDX9aoFwowjcdSC4E8w=="],
+
     "@napi-rs/wasm-runtime": ["@napi-rs/wasm-runtime@0.2.4", "", { "dependencies": { "@emnapi/core": "^1.1.0", "@emnapi/runtime": "^1.1.0", "@tybys/wasm-util": "^0.9.0" } }, "sha512-9zESzOO5aDByvhIAsOy9TbpZ0Ur2AJbUI7UT73kcUTS2mxAMHOBaa1st/jAymNoCtvrit99kkzT1FZuXVcgfIQ=="],
 
     "@nestjs/apollo": ["@nestjs/apollo@13.1.0", "", { "dependencies": { "@apollo/server-plugin-landing-page-graphql-playground": "4.0.1", "iterall": "1.3.0", "lodash.omit": "4.5.0", "tslib": "2.8.1" }, "peerDependencies": { "@apollo/gateway": "^2.0.0", "@apollo/server": "^4.11.3", "@apollo/subgraph": "^2.0.0", "@as-integrations/fastify": "^2.1.1", "@nestjs/common": "^11.0.1", "@nestjs/core": "^11.0.1", "@nestjs/graphql": "^13.0.0", "graphql": "^16.10.0" }, "optionalPeers": ["@apollo/gateway", "@apollo/subgraph", "@as-integrations/fastify"] }, "sha512-/FRg195AxpZ58Kjd7geeksaRs3blmlGMDUak7WGbrl7ZWX7J9VuulhTjVHP6Z+dhH4Tn+AsyjkkJ2euC8psv3A=="],
@@ -631,6 +633,12 @@
     "@octokit/rest": ["@octokit/rest@20.1.2", "", { "dependencies": { "@octokit/core": "^5.0.2", "@octokit/plugin-paginate-rest": "11.4.4-cjs.2", "@octokit/plugin-request-log": "^4.0.0", "@octokit/plugin-rest-endpoint-methods": "13.3.2-cjs.1" } }, "sha512-GmYiltypkHHtihFwPRxlaorG5R9VAHuk/vbszVoRTGXnAsY60wYLkh/E2XiFmdZmqrisw+9FaazS1i5SbdWYgA=="],
 
     "@octokit/types": ["@octokit/types@13.10.0", "", { "dependencies": { "@octokit/openapi-types": "^24.2.0" } }, "sha512-ifLaO34EbbPj0Xgro4G5lP5asESjwHracYJvVaPIyXMuiuXLlhic3S47cBdTb+jfODkTE5YtGCLt3Ay3+J97sA=="],
+
+    "@open-draft/deferred-promise": ["@open-draft/deferred-promise@2.2.0", "", {}, "sha512-CecwLWx3rhxVQF6V4bAgPS5t+So2sTbPgAzafKkVizyi7tlwpcFpdFqq+wqF2OwNBmqFuu6tOyouTuxgpMfzmA=="],
+
+    "@open-draft/logger": ["@open-draft/logger@0.3.0", "", { "dependencies": { "is-node-process": "^1.2.0", "outvariant": "^1.4.0" } }, "sha512-X2g45fzhxH238HKO4xbSr7+wBS8Fvw6ixhTDuvLd5mqh6bJJCFAPwU9mPDxbcrRtfxv4u5IHCEH77BmxvXmmxQ=="],
+
+    "@open-draft/until": ["@open-draft/until@2.1.0", "", {}, "sha512-U69T3ItWHvLwGg5eJ0n3I62nWuE6ilHlmz7zM0npLBRvPRd7e6NYmg54vvRtP5mZG7kZqZCFVdsTWo7BPtBujg=="],
 
     "@opensearch-project/opensearch": ["@opensearch-project/opensearch@3.5.1", "", { "dependencies": { "aws4": "^1.11.0", "debug": "^4.3.1", "hpagent": "^1.2.0", "json11": "^2.0.0", "ms": "^2.1.3", "secure-json-parse": "^2.4.0" } }, "sha512-6bf+HcuERzAtHZxrm6phjref54ABse39BpkDie/YO3AUFMCBrb3SK5okKSdT5n3+nDRuEEQLhQCl0RQV3s1qpA=="],
 
@@ -1770,6 +1778,8 @@
 
     "is-negative-zero": ["is-negative-zero@2.0.3", "", {}, "sha512-5KoIu2Ngpyek75jXodFvnafB6DJgr3u8uuK0LEZJjrU19DrMD3EVERaR8sjz8CCGgpZvxPl9SuE1GMVPFHx1mw=="],
 
+    "is-node-process": ["is-node-process@1.2.0", "", {}, "sha512-Vg4o6/fqPxIjtxgUH5QLJhwZ7gW5diGCVlXpuUfELC62CuxM1iHcRe51f2W1FDy04Ai4KJkagKjx3XaqyfRKXw=="],
+
     "is-number": ["is-number@7.0.0", "", {}, "sha512-41Cifkg6e8TylSpdtTpeLVMqvSBEVzTttHvERD741+pnZ8ANv0004MRL43QKPDlK9cGvNp6NZWZUBlbGXYxxng=="],
 
     "is-number-object": ["is-number-object@1.1.1", "", { "dependencies": { "call-bound": "^1.0.3", "has-tostringtag": "^1.0.2" } }, "sha512-lZhclumE1G6VYD8VHe35wFaIif+CTy5SJIi5+3y4psDgWu4wPDoBhF8NxUOinEc7pHgiTsT6MaBb92rKhhD+Xw=="],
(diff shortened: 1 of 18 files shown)


michaelbromley:
## Pre-merge 5 — stripe@13.11.0 vs stripe@22.2.0 d.ts diff for fields the plugin reads/writes

Pulled `stripe@13.11.0` from npm and diffed the d.ts files in `types/` against `stripe@22.2.0`'s `cjs/resources/` for the exact fields the plugin touches. Result: **shape-identical for everything the plugin reads or writes**.

| Field | v13.11.0 | v22.2.0 | Same? |
|-------|----------|---------|-------|
| `PaymentIntent.id` | `string` | `string` | ✅ |
| `PaymentIntent.client_secret` | `string \| null` | `string \| null` | ✅ |
| `PaymentIntent.amount_received` | `number` | `number` | ✅ |
| `PaymentIntent.status` | `PaymentIntent.Status` | `PaymentIntent.Status` | ✅ |
| `PaymentIntent.metadata` | `Stripe.Metadata` (= `Record<string, string>`) | `Metadata` (= `Record<string, string>`) | ✅ |
| `PaymentIntent.last_payment_error` | `PaymentIntent.LastPaymentError \| null` | `PaymentIntent.LastPaymentError \| null` | ✅ |
| `PaymentIntent.automatic_payment_methods` | `PaymentIntent.AutomaticPaymentMethods \| null` | `PaymentIntent.AutomaticPaymentMethods \| null` | ✅ |
| `PaymentIntentCreateParams.automatic_payment_methods.enabled` | `boolean` | `boolean` | ✅ |
| `Refund.status` | `string \| null` | `string \| null` | ✅ |
| `Refund.failure_reason` | `string?` | `string?` | ✅ |
| `RefundCreateParams.reason` | `'duplicate' \| 'fraudulent' \| 'requested_by_customer' \| 'expired_uncaptured_charge'` | `'duplicate' \| 'fraudulent' \| 'requested_by_customer'` | ⚠️ narrowed |
| `CustomerCreateParams.email` | `string?` | `string?` | ✅ |
| `CustomerCreateParams.name` | `string?` | `string?` | ✅ |
| `Charge.refunded` | `boolean` | `boolean` | ✅ (not currently read) |
| `Charge.amount_refunded` | `number` | `number` | ✅ (not currently read) |

### The only real change worth a note

`RefundCreateParams.reason` dropped `'expired_uncaptured_charge'` between v13 and v22. The plugin only sends one of the surviving three (`requested_by_customer`, `duplicate`, `fraudulent`) — anything else falls through to `metadata.vendureRefundReason`. So nothing actually exercised the dropped value, and the tripwire in `src/__type-checks.ts` pins this so a future SDK change would break the build, not the runtime.

(`Refund.Reason` *responses* still include `'expired_uncaptured_charge'` in v22 — the narrow only applies to what you can pass on create.)

### Conclusion

The "default API version moves 2023-08-16 → 2026-05-27.dahlia" risk listed in the confidence plan is materially smaller than feared: the type shapes the plugin reads/writes are unchanged across the entire SDK and API-version span, modulo one narrowed enum the plugin doesn't use.

LeftoversTodayAppAdmin:
@michaelbromley +1 to updating the Stripe SDK!

PR #38, opened 2026-06-12
fix(stripe-plugin): pass requestOptions to customer list/create calls

### LINKED ISSUE #25: Stripe Plugin: storeCustomersInStripe ignores requestOptions, causing customer to be created in wrong account context
## Describe the bug

When using the Stripe plugin with Stripe Connect and connected-account payments via `requestOptions`, checkout can fail after account creation / login with:

`No such customer: 'cus_...'`

On the storefront side, the Stripe Payment Element does not render because `createStripePaymentIntent` fails before returning a client secret.

The referenced code is in `stripe.service.ts`, specifically in `getStripeCustomerId()`.

## Root cause

`createPaymentIntent()` already resolves and passes `requestOptions` into `stripe.paymentIntents.create(...)`, and the [docs](https://docs.vendure.io/current/community-plugins/stripe-plugin/stripe-plugin-options#requestoptions) show that we can include an `accountId` for use with Connect accounts, so PaymentIntents are created in the connected account context.

However, `getStripeCustomerId()` does not pass those same `requestOptions` into Stripe customer lookup / creation.

The relevant local patch was:

```ts
const requestOptions = await this.options.requestOptions?.(
  new Injector(this.moduleRef),
  ctx,
  order,
);
```

and then using that `requestOptions` in the Stripe customer calls.

Without that, the flow becomes:

* `paymentIntents.create(...)` → connected account
* `customers.list(...)` / `customers.create(...)` → platform account

This creates an account-scoping mismatch. A Stripe customer created or looked up in the platform account may later be reused in the connected account, where Stripe correctly returns:

`No such customer`

## Expected behavior

When `requestOptions` is configured, Stripe customer lookup / creation should happen in the same Stripe account context as PaymentIntent creation.

For example, if PaymentIntents are created with:

```ts
{ stripeAccount: connectedAccountId }
```

then customer lookup / creation should use the same request options.

## Current behavior

In our case, guest checkout worked, but after registration / login the checkout flow started failing because the stored or resolved Stripe customer belonged to a different Stripe account context than the PaymentIntent creation call.

## Local fix tested

I patched `getStripeCustomerId()` so it resolves `requestOptions` and passes them into:

* `stripe.customers.list(...)`
* `stripe.customers.create(...)`

After this patch, customer lookup / creation happened in the same connected-account context as PaymentIntent creation, and the flow worked correctly.

## Suggested upstream fix

Inside `getStripeCustomerId()`:

1. resolve `requestOptions` the same way `createPaymentIntent()` already does
2. pass them into Stripe customer lookup / creation calls

---

Notes:

- Stripe restricted-keys cause permission error for customer access on the connected account, which disappeared once the appropriate permissions were granted.
- `storeCustomersInStripe` is not necessarily the root cause by itself, but it makes the missing `requestOptions` in `getStripeCustomerId()` visible and breaking.

BraynioAndy:
## Problem

With `storeCustomersInStripe: true` and a `requestOptions` callback that scopes calls to a connected account (`stripeAccount`), `createPaymentIntent()` passes the options to `paymentIntents.create()`, but `getStripeCustomerId()` does not pass them to `customers.list()` / `customers.create()`. The customer is looked up/created on the **platform** account while the PaymentIntent targets the **connected** account, so intent creation fails with `No such customer: 'cus_...'`.

The platform-scoped id is then cached on `customer.customFields.stripeCustomerId`, so every subsequent attempt fails as well — manually clearing the field doesn't help, because the next lookup re-fetches the same platform customer by email and writes it back.

## Fix

Resolve `requestOptions` in `getStripeCustomerId()` the same way `createPaymentIntent()` does and pass them to both customer calls. The options hash is only passed when non-empty, because the Stripe SDK rejects an empty options object (`isOptionsHash` requires at least one known key).

## Testing

- New e2e test asserting that both customer calls carry the `Stripe-Account` header (red without the fix — nock doesn't match the requests).
- Full stripe-plugin e2e suite passes (19/19); lint introduces no new warnings.
- Verified end-to-end against a real Stripe sandbox using Connect: logged-in customer, Payment Element renders, payment settles via webhook on the connected account.

Fixes #25

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/38"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783870087&installation_id=128408429&pr_number=38&repository=vendurehq%2Fcommunity-plugins&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fcommunity-plugins%2Fpull%2F38&signature=3a22a880fa7fb4849b998a289925ab8f69bc6110fc00ad4ae6aa7d129fd914d6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/stripe-plugin/e2e/stripe-payment.e2e-spec.ts
+++ b/packages/stripe-plugin/e2e/stripe-payment.e2e-spec.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { CurrencyCode, LanguageCode } from '@vendure/common/lib/generated-types';
-import { EntityHydrator, mergeConfig } from '@vendure/core';
+import { Customer, EntityHydrator, mergeConfig, TransactionalConnection } from '@vendure/core';
 import {
     createProductDocument,
     createProductVariantsDocument,
@@ -326,6 +326,50 @@ describe('Stripe payments', () => {
             description: `Description for ${activeOrder!.customer!.emailAddress}`,
             phone: '12345',
         });
+    });
+
+    // https://github.com/vendurehq/community-plugins/issues/25
+    it('should pass requestOptions to customer list and create calls', async () => {
+        StripePlugin.options.requestOptions = () => ({
+            stripeAccount: 'acct_connected',
+        });
+
+        // Reset the cached Stripe customer id so the lookup/create path runs again.
+        const connection = server.app.get(TransactionalConnection);
+        const customerRepo = connection.rawConnection.getRepository(Customer);
+        const dbCustomer = await customerRepo.findOneByOrFail({
+            emailAddress: customers[1].emailAddress,
+        });
+        dbCustomer.customFields.stripeCustomerId = '';
+        await customerRepo.save(dbCustomer, { reload: false });
+
+        await shopClient.asUserWithCredentials(customers[1].emailAddress, 'test');
+
+        // Both customer calls must carry the Stripe-Account header, otherwise the
+        // customer is created on the platform account while the payment intent
+        // targets the connected account ("No such customer" on intent creation).
+        const stripeAccountHeaders: string[] = [];
+        nock('https://api.stripe.com/', {
+            reqheaders: {
+                'Stripe-Account': headerValue => {
+                    stripeAccountHeaders.push(headerValue);
+                    return true;
+                },
+            },
+        })
+            .get(/\/v1\/customers.*/)
+            .reply(200, { data: [] })
+            .post('/v1/customers')
+            .reply(201, { id: 'connected-account-customer-id' });
+        nock('https://api.stripe.com/').post('/v1/payment_intents').reply(200, {
+            client_secret: 'test-client-secret',
+        });
+
+        const { createStripePaymentIntent } = await shopClient.query(createStripePaymentIntentDocument);
+        expect(stripeAccountHeaders).toEqual(['acct_connected', 'acct_connected']);
+        expect(createStripePaymentIntent).toEqual('test-client-secret');
+        StripePlugin.options.requestOptions = undefined;
+        StripePlugin.options.customerCreateParams = undefined;
     });
 
     // https://github.com/vendurehq/vendure/issues/2450
--- a/packages/stripe-plugin/e2e/stripe-payment.e2e-spec.ts
+++ b/packages/stripe-plugin/e2e/stripe-payment.e2e-spec.ts
@@ -278,7 +278,7 @@ describe('Stripe payments', () => {
         });
         expect(connectedAccountHeader).toEqual('acct_connected');
         expect(createStripePaymentIntent).toEqual('test-client-secret');
-        StripePlugin.options.paymentIntentCreateParams = undefined;
+        StripePlugin.options.requestOptions = undefined;
     });
 
     // https://github.com/vendurehq/vendure/issues/2412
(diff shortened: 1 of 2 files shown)


BraynioAndy:
I have read the CLA Document and I hereby sign the CLA

BraynioAndy:
recheck

michaelbromley (changes requested):
Thanks for tracking this down — the connected-account mismatch is a real issue and the fix is on the right track. A few things before merge:

1. `requestOptions` gets resolved twice per `createPaymentIntent` call when `storeCustomersInStripe: true` — once in `createPaymentIntent` and again in `getStripeCustomerId`. If someone's callback does any real work (DB lookup for the connected account, etc.) that's wasted. Worth extracting a small `resolveRequestOptions(ctx, order)` helper and calling it from both places. That also kills the `opts ? call(x, opts) : call(x)` repetition around `customers.list` / `customers.create`.

2. There's a pre-existing cleanup bug in `stripe-payment.e2e-spec.ts:281` — it resets `paymentIntentCreateParams` where it meant to reset `requestOptions`. Your new test happens to set `requestOptions` to the same value so it works, but it's masking the leak. Since you're already in this file, worth fixing that line too.

3. The new comment in `getStripeCustomerId` mostly narrates the diff ("resolve the same way as createPaymentIntent does..."). Could you swap it for one that explains *why* — customer ops need to hit the same connected account as the intent, and the SDK rejects an empty options object.

4. Minor: `stripeCustomerId = ''` in the new test — the field is `nullable: true`, `null` reads more honestly.

BraynioAndy:
Thanks for the thorough review, @michaelbromley! Pushed a follow-up commit addressing all four points:

1. **Double resolution** — extracted a `resolveRequestOptions(ctx, order)` helper. `createPaymentIntent` now resolves the options *once* and passes them into `getStripeCustomerId(ctx, order, requestOptions)`, so the callback no longer runs twice per request when `storeCustomersInStripe` is enabled. The resolved value (or `undefined`) is passed straight to `customers.list`/`customers.create`, which also removed the `opts ? call(x, opts) : call(x)` branches.
2. **Cleanup leak** — `stripe-payment.e2e-spec.ts` now resets `requestOptions` instead of `paymentIntentCreateParams` in the requestOptions test.
3. **Comment** — rewrote the one in `getStripeCustomerId` to explain *why* (customer ops must hit the same connected account as the intent; empty options hash rejected by the SDK) rather than narrate the diff.
4. **Minor** — the new test now clears `stripeCustomerId` with `null` instead of `''`.

Build, lint and the e2e suite (19/19) pass locally. Ready for another look when you have a moment.

michaelbromley (approved):

PR #41, opened 2026-07-06
fix(elasticsearch-plugin): import health-check types from @vendure/core

tbouliere-datasolution:
Description :
## Problem

`ElasticsearchPlugin` crashes at module load with `Cannot find module '@nestjs/terminus'` on Vendure 3.7.

Vendure 3.7 removed its transitive dependency on `@nestjs/terminus`, replacing the terminus types with an in-core compatibility shim (`@vendure/core` `health-check/terminus-compat`). `elasticsearch.health.ts` still imported `HealthCheckError`, `HealthIndicator`, and `HealthIndicatorResult` directly from `@nestjs/terminus`, and the package declared terminus in neither `dependencies` nor `peerDependencies` — it only ever resolved transitively through `@vendure/core`. As a result the plugin no longer loads.

## Fix

Import those three symbols from `@vendure/core` instead of `@nestjs/terminus`, as recommended by the `terminus-compat` shim's own documentation ("migrate by changing the import path to `@vendure/core`"). The shim types are structurally compatible, so no other code change is needed.

## Version bump

The compat shim was verified to be **absent** in `3.6.0` / `3.6.4` and **present from `3.7.0`**. Because the new import only resolves against `@vendure/core >= 3.7.0`, this bumps:

- `@vendure/core` and `@vendure/common` peer ranges: `^3.6.0-0` → `^3.7.0`
- `@vendure/{core,common,testing}` dev dependencies: `3.6.0-minor-202603280303` → `3.7.0`

This raises the plugin's minimum supported Vendure version to `3.7.0`.

## Notes

- `plugin.ts` already imports `HealthCheckRegistryService` from `@vendure/core`; this makes the health file consistent with the rest of the plugin.
- The `ElasticsearchHealthIndicator` is already `@deprecated` (slated for removal in v4.0.0); the shim surfaces a deprecation hint on `HealthIndicatorResult`, which is non-fatal and expected.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/41"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785938570&installation_id=128408429&pr_number=41&repository=vendurehq%2Fcommunity-plugins&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fcommunity-plugins%2Fpull%2F41&signature=0bbae4c0ce48b404429e25aea6749d41c3188fd516760163f9994db3bf4fdf14"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -55,8 +55,9 @@
     },
     "packages/braintree-plugin": {
       "name": "@vendure-community/braintree-plugin",
-      "version": "1.0.0",
+      "version": "1.0.1",
       "devDependencies": {
+        "@nestjs/testing": "^11.1.13",
         "@types/braintree": "^3.3.11",
         "@vendure/common": "3.6.0-minor-202603280303",
         "@vendure/core": "3.6.0-minor-202603280303",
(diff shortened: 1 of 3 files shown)


michaelbromley:
Hi, thanks for bringing this up and proposing a solution. I’m going to go with a slightly simpler and less invasive solution of just completely removing the health check code altogether, because it is not even used anyway, and bumping the Vendure required version for types that don’t actually do anything at runtime is not ideal.

PR #48, opened 2026-08-04
fix(elasticsearch-plugin): don't strand temp index or abort bootstrap on create failure

### LINKED ISSUE #47: [elasticsearch-plugin] Failed temp index creation strands the index and aborts application bootstrap
## Summary

A failure while creating the throwaway index used by the startup settings/mappings drift check has two compounding consequences: the temp index is leaked into the cluster permanently, and the error propagates out of `onApplicationBootstrap()` and aborts Nest bootstrap. Three separate places in the package combine to produce this; each is arguably defensible alone, but together they turn a transient cluster hiccup into a stranded index plus a half-initialised application.

All line references are against `d0847a14a2609c7a9a27f7530b9f8fa853fe40a5` (current `main`).

## The chain

**1. `createIndices()` swallows the failure instead of reporting it.**

https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/indexing/indexing-utils.ts#L98-L105

The `catch` logs and does not rethrow, so `createIndices()` resolves normally whether or not the index was actually created. Every caller proceeds as if it exists.

The diagnostic that replaced the throw also loses the information. `message` and `stack` are non-enumerable on the client error classes, so `JSON.stringify(e, null, 2)` drops both. For a `TimeoutError` — the failure mode below — the entire logged line is:

```json
{ "name": "TimeoutError", "meta": {} }
```

(verified against `@opensearch-project/opensearch` 3.5.1 and `@elastic/elasticsearch` 9.3.4; for a `ResponseError` you get `name` plus the raw `meta`, still without the message).

**2. The temp index is deleted as a bare statement, with no `try`/`finally`.**

https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/elasticsearch.service.ts#L133-L199

Because step 1 does not throw, execution continues to `getSettings({ index: tempIndex })` on an index that was never created, which raises `index_not_found_exception`. The delete at the end of the block is never reached, so the temp index — if the server did create it, e.g. the request succeeded server-side but timed out client-side — is stranded. There is no later cleanup path for `temp-*` indices.

**3. The call is unguarded in `onApplicationBootstrap()`.**

https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/plugin.ts#L123-L134

The error from step 2 propagates out of the unguarded `await` and aborts bootstrap. The asymmetry is what marks this as unintended rather than a deliberate fail-fast: `checkConnection()` immediately above **is** wrapped and degrades gracefully with a `return`, so the plugin already treats "search backend unavailable" as non-fatal. Only this one call is unprotected.

### Latent, not currently firing

Line 198 deletes `tempPrefix + 'variants'` (hardcoded), while the temp index is bound at line 136 as `tempPrefix + indexName`. These agree **only** because `VARIANT_INDEX_NAME === 'variants'` and `createIndex()` is called once with that constant. It is a latent bug, not a live one — it becomes live the moment a second index name is introduced. Worth fixing alongside the rest, but it is not part of the failure described above.

### Related, same swallow pattern

`deleteIndices()` ([L109-L118](https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/indexing/indexing-utils.ts#L109-L118)) and `deleteByChannel()` ([L121-L135](https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/indexing/indexing-utils.ts#L121-L135)) log-and-continue identically, and additionally pass the error object where `Logger.error()` expects a string.

Note also that `indexer.controller.ts` [L246-L259](https://github.com/vendurehq/community-plugins/blob/d0847a14a2609c7a9a27f7530b9f8fa853fe40a5/packages/elasticsearch-plugin/src/indexing/indexer.controller.ts#L246-L259) already wraps `createIndices()` in a `try`/`catch` that logs `'Could not recreate indices.'` and rethrows. That handler is currently unreachable, because `createIndices()` never throws — good evidence that throwing was the intended contract.

## Reproduction

Any condition that makes the temp-index `PUT` fail or time out while the live index already exists:

1. Configure the plugin against a reachable cluster and let it create the `variants` index.
2. Set a client request timeout shorter than the temp-index create takes — e.g. `requestTimeout: 500` in the client options, or point the client at a cluster under load. A large `indexSettings` payload (custom analyzers/synonym filters) makes this easy to hit at realistic timeouts.
3. Restart the application.

Observed: a log line with no usable message, `index_not_found_exception` propagating out of bootstrap, and a `temp-<epochMs>-<salt>-variants` index left in the cluster. Each restart leaks another one.

## Impact

We hit this in production on Vendure 3.7.1 with OpenSearch 3.1 on AWS. The client request timeout was 8s; the drift check's `PUT /temp-...-variants` carried a ~6.4KB body (mappings plus a `synonym_graph` filter of 62 groups) and exceeded it.

The consequential part was the failure mode on the **worker**: it died mid-bootstrap but did not exit. Its scheduler timers were already armed, so cron tasks kept firing and the process looked healthy to our health checks, while `startJobQueue()` was never reached and no queued job was consumed for 3.5 hours. Orders sat in `PaymentSettled` because the queue that advances them had no consumer.

Separately, CloudWatch showed `Shards.active` climbing steadily while document count stayed flat — the accumulated stranded temp indices.

## Suggested fix

The guard belongs inside the service around the diagnostic branch, not around the whole `createIndicesIfNotExists()` call in `plugin.ts` — wrapping the outer call would also swallow a genuine live-index creation failure. Concretely:

1. `createIndices()` rethrows after logging, with error rendering that does not lose `message`.
2. The `if (!result.body)` branch (the one creating the **real** index) gets its own `catch` that logs and continues, preserving today's boot semantics — without this, (1) would convert "index missing, no search" into "every instance crash-loops".
3. The drift-check branch gets `try`/`catch` → `Logger.warn` plus a `finally` that always deletes the computed `tempIndex` (also removing the hardcoded `'variants'`), with the delete's own failure guarded so it cannot mask the original error.

I have a PR ready implementing exactly this, with unit tests — will link it below.

brmk:
Fixes #47.

## Problem

When the startup drift check failed to create its throwaway index, two things happened: the temp index was leaked into the cluster permanently, and the resulting error aborted Nest bootstrap.

The chain, all on current `main`:

1. `createIndices()` caught the failure, logged it, and **returned normally**. Callers proceeded as if the index existed.
2. Execution therefore reached `getSettings({ index: tempIndex })` on an index that was never created → `index_not_found_exception`.
3. That propagated out of the unguarded `await this.elasticsearchService.createIndicesIfNotExists()` in `onApplicationBootstrap()` and aborted bootstrap.
4. The temp index delete was a bare statement at the end of the block, so it was never reached — if the server had in fact created the index (request succeeded server-side, timed out client-side), it stayed forever. There is no later `temp-*` cleanup path.

The issue has the full write-up, including the production failure mode: a worker that died mid-bootstrap but did not exit, so cron kept firing while `startJobQueue()` was never reached and the job queue had no consumer.

## Approach

The three edits are ordered deliberately, and **item 2 is what makes item 1 safe**.

**1. `createIndices()` rethrows after logging.**

An index that was not created is not something a caller can paper over. This also makes reachable the `catch` in `indexer.controller.ts` that already logs `'Could not recreate indices.'` and rethrows — dead code today, and good evidence that throwing was the intended contract.

**2. The branch that creates the *real live* index gets its own `catch` that logs and continues.**

This deliberately preserves today's boot semantics exactly. Without it, item 1 would convert the current behaviour ("index missing, this instance serves no search results") into "every instance crash-loops, whole shop down" — an unacceptable regression, and the reason this isn't just a one-line rethrow.

**3. The drift-check branch gets `try`/`catch` → `Logger.warn`, plus a `finally` that always deletes the computed `tempIndex`.**

This branch is *purely diagnostic* — it builds a throwaway index from the current config and compares it to the live one to warn about drift. It should never be able to abort bootstrap, and it must always clean up after itself.

Two details worth calling out:

- The delete is attempted **even when the create failed**, because a client-side timeout does not prove the server did not create the index — that is exactly the case that was stranding indices.
- The delete's own failure is caught and warned, so it cannot mask the error that brought us out of the `try`.

This also drops the hardcoded `tempPrefix + 'variants'` in favour of the computed `tempIndex`. That one was **latent, not live**: the two agreed only because `VARIANT_INDEX_NAME === 'variants'` and `createIndex()` is called once with that constant.

### Why not just wrap the call in `plugin.ts`?

A reviewer will reasonably ask why item 3 isn't a one-liner `try`/`catch` around `createIndicesIfNotExists()` in `onApplicationBootstrap()`. Because that would also swallow a genuine **live**-index creation failure and every other error in the method, collapsing "the drift check couldn't run" and "the search index is broken" into one silent path. The guard belongs inside the service, scoped to the diagnostic branch only, so the two cases stay distinguishable in the logs.

### Error logging

`JSON.stringify()` is the wrong tool for these errors: `message` and `stack` are non-enumerable on the ES/OS client error classes, so serializing drops them. A `TimeoutError` — the exact failure in the linked issue — logged as:

```json
{ "name": "TimeoutError", "meta": {} }
```

`describeSearchClientError()` pulls out `message` and, when present, the server response body from `meta.body`. Verified against `@opensearch-project/opensearch` 3.5.1 and `@elastic/elasticsearch` 9.3.4, and covered by a test that constructs the real client error classes rather than a hand-rolled stand-in.

## Tests

Added `src/indexing/indexing-utils.spec.ts` and `src/elasticsearch.service.spec.ts` (unit only — no e2e, no Docker), covering:

- `createIndices()` rethrows, and logs a message that survives serialization
- a live-index create failure does **not** abort bootstrap
- the temp index is deleted using the name it was created with (regression guard for the hardcoded `'variants'`)
- the temp index is deleted when the drift check throws
- cleanup is still attempted when the temp index create itself failed
- a failing cleanup does not mask the original error

I checked these are not vacuous: reverting the two source files and re-running leaves 8 of the new assertions failing.

## Gates

Run with Bun, per `CONTRIBUTING.md`. No lockfile changes are included in this PR.

| Command | Scope | Result |
| --- | --- | --- |
| `bun install` | root | clean |
| `bun run build` | root (lerna, 8 projects) | pass |
| `bun run lint` | root (lerna, 8 projects) | pass — 0 errors |
| `bun run test` | root (lerna, 3 projects) | pass — 50 tests |
| `bun run test` | elasticsearch-plugin | pass — 45 tests |

Lint reports 12 pre-existing warnings in the package (`no-misused-promises`, `require-await`, `import/order`); none are in the files or lines touched here, and no new warnings were introduced.

`bun run e2e` was **not** run — it requires Docker, which was out of scope for this change.

## Notes for maintainers

- `deleteIndices()` and `deleteByChannel()` in the same file have the identical log-and-continue pattern, and additionally pass an error object where `Logger.error()` expects a string. I left both alone to keep this PR scoped to the bootstrap failure; noted in the issue if you'd like them addressed separately.
- `plugin.ts` is intentionally untouched.

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788447975&installation_model_id=20193&pr_number=48&repository=vendurehq%2Fcommunity-plugins&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fcommunity-plugins%2Fpull%2F48&signature=2e0d3fbb80e90f62526044a8c820d40e45fcaa6f23174766508f88dd0ebf0213"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld:
Thanks for this, @brmk. We've picked it up in our maintainer lane.

For pull requests in our maintainer lane, our standard process is to independently verify
the change, push any required follow-up commits to the contributor's branch, and merge it
after the required evidence and checks pass. This avoids an extra review round trip.

What we changed:

- `createIndices()` no longer logs the failure it rethrows, so each failure is logged once, by the caller that handles it.
- The reindex error path now uses `describeSearchClientError()` and names the index, instead of `JSON.stringify(e)`.
- Pinned the service's own message in the spec, so the assertion can no longer be satisfied by the log inside `createIndices()`.
- Added a test that a failed drift check warns rather than errors.
- Renamed the temp index cleanup test, which passes against the pre-fix code and so does not guard the name.

Verification: unit 46 passed; e2e 99 passed on Elasticsearch 9.1 and 99 passed on OpenSearch 3.0.
Deleting the service's own error log fails the pinned assertion.

biggamesmallworld (approved):
Verified locally: reproduced the bootstrap abort on the pre-fix code, confirmed the server boots and warns once with the fix. Unit 46 passed, e2e 99 passed on Elasticsearch 9.1 and 99 on OpenSearch 3.0.

PR #49, opened 2026-08-05
fix(mollie-plugin): remove false positive error logs `... should be refunded`

martijnvdbrug:
Removes the misleading error log that was shown when a Mollie webhook arrived for a cancelled order.

This is what happened:

1. Order is cancelled in Vendure
2. Vendure does a refund with Mollie
3. Mollie sends a webhook to Vendure with the **payment + refund**, where the payment is still `paid`
4. Order is cancelled in Vendure, so the log incorrectly says a cancelled order was paid for.

**Error log for duplicate payments for the same order is still in place!**

---
<a href="https://app.blacksmith.sh/vendurehq/codesmith/community-plugins/pr/49"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788513130&installation_model_id=20193&pr_number=49&repository=vendurehq%2Fcommunity-plugins&return_to=https%3A%2F%2Fgithub.com%2Fvendurehq%2Fcommunity-plugins%2Fpull%2F49&signature=7063b916381824098d9bdd73b2d4a543c0bdc986c46028a0196b4d3c683e8509"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


biggamesmallworld:
Thanks @martijnvdbrug, the false positive you describe is real and worth removing.

One case gets lost with the branch deleted outright. An order cancelled before it was ever
placed has no `orderPlacedAt`, so the duplicate-payment check above is skipped, and
`Cancelled` is not in `VENDURE_STATES_THAT_REQUIRE_ACTION`. If the customer then completes
the Mollie payment, the merchant holds the money against a cancelled order and the only
trace is the info log "no need for handling".

What separates that from your false positive is whether the Mollie payment is already
recorded on the order. Keeping the error only when it is not covers both:

```ts
const paymentWithSameTransactionId = order.payments.find(p => p.transactionId === molliePayment.id);
```

That is the same lookup the duplicate-payment branch does, so it can be shared.

Worth adding two e2e cases, since neither behaviour is covered today: a cancelled order with
a matching recorded payment logs nothing, and a cancelled order without one still logs the
error. We tried this locally and both hold: mollie e2e 31 passed, and each test fails if the
corresponding behaviour is removed.

One more thing before this merges: the commit subject is truncated mid-word, and the
changelog is generated from these commits.

Happy to push the change to your branch if you would rather not.

martijnvdbrug:
@biggamesmallworld oh wow, good catch! Feel free to add the changes to this pr if you already have them locally.  

e2e are in place here indeed, but i am a little short on time right now. Might take a couple of weeks before I get to it.

PR #51, opened 2026-08-09
feat(elasticsearch-plugin): avoid redundant reindex on stock movements

### LINKED ISSUE #50: elasticsearch-plugin: stock movements that dont change inStock trigger a full reindex that briefly drops the product from search
## Summary

On every `StockMovementEvent` the plugin re-indexes the affected product by deleting its documents and then recreating them. This causes two problems.

**1. The product briefly disappears from search.** The delete and the recreate happen as separate bulk operations (each with `refresh`), and the update runs as a queued job that is picked up asynchronously. So for a window, in our testing often a few hundred milliseconds, a search that should return the product returns nothing. On a storefront this is a visible bug: an in-stock product flickers out of the results and back in whenever its stock changes, which happens constantly during ordering.

**2. It is wasteful.** A stock change that does not alter any indexed field (for example 50 to 49, or any movement at all once stock is no longer stored in the index document) still rewrites a byte-for-byte identical document. On a busy catalog that is a lot of unnecessary search-engine writes, job-queue load, and replication traffic, all producing no change in the index.

## Where it happens

- `packages/elasticsearch-plugin/src/plugin.ts`: the `StockMovementEvent` subscriber enqueues an `update-variants` job for every movement, with no check for whether anything relevant changed.
- `packages/elasticsearch-plugin/src/indexing/indexer.controller.ts`: `updateProductsOperations` calls `deleteProductOperations` and then `updateProductsOperationsOnly`, so a product's documents are removed and rebuilt on every update, again with no diff.

## Proposal

Two opt-in options, both defaulting to the current behavior so existing installs are unaffected.

**`reindexOnStockMovement: 'always' | 'onStockStatusChange'`** (default `'always'`). In `'onStockStatusChange'` mode the `StockMovementEvent` subscriber only enqueues an update when the movement actually flips a variant's `inStock` or its product's `productInStock`. A 50 to 49 change never touches search, so no job is created at all. This inspects only the built-in stock booleans, so it is meant for setups that do not map a stock-derived custom field. Those should stay on `'always'` and use the second option.

**`skipUnchangedIndexUpdates: boolean`** (default `false`). Before the delete-then-recreate, the indexer builds the product's target documents, compares them against what is currently indexed, and skips the write when they are identical. Because it compares the whole document it stays correct for any mapping configuration, including stock-derived custom mappings, and it removes the flicker for any no-op update regardless of what triggered it. A full reindex is never skipped.

The two are complementary. The first avoids even creating a job for stock movements that cannot change search results. The second is a general safety net that prevents a redundant delete-then-recreate from any trigger.

## Backwards compatibility

- Both options default to today's behavior, so upgrading and changing nothing behaves exactly as before.
- No changes to the `SearchClientAdapter` interface, so custom adapters keep working. The comparison read goes through the existing `search` method.
- No index mapping or schema changes.
- The stock-status check uses Vendure's `ProductVariantService.getSaleableStockLevel`, and the document comparison goes through the adapter, so both features behave identically on Elasticsearch and OpenSearch.

## Impact if resolved

- A stock change that does not affect search membership no longer removes the product from the index even momentarily, so the flicker is gone.
- Ordinary stock churn stops generating redundant writes and reindex jobs.

I have a PR ready that implements both options, with unit tests for the document comparison and e2e tests for the behavior on both backends. Happy to adjust the option names or defaults to your preference.

LeftoversTodayAppAdmin:
Closes #50.

Adds two opt-in options that stop stock movements from needlessly re-indexing a product, including the delete-then-recreate window during which the product drops out of search results. Both default to the current behavior, so existing installs are unaffected.

### `reindexOnStockMovement: 'always' | 'onStockStatusChange'` (default `'always'`)

Pre-enqueue guard in the `StockMovementEvent` subscriber. In `'onStockStatusChange'` it only enqueues an update when the movement flips a variant's `inStock` or its product's `productInStock`, so a 50 to 49 change never creates a job. It inspects only the built-in stock booleans, so it is meant for setups without a stock-derived custom mapping. Others should stay on `'always'` and use the option below.

### `skipUnchangedIndexUpdates: boolean` (default `false`)

Write-path guard. Before the delete-then-recreate, `updateProductsOperations` builds the product's target documents, compares them against what is currently indexed, and skips the write when they are identical. It compares the whole document, so it stays correct for any mapping configuration including stock-derived custom mappings, and it removes the flicker for any no-op update regardless of trigger. A full reindex is never skipped.

The two are complementary: the first avoids creating a job for stock movements that cannot change search results, the second is a general safety net against a redundant delete-then-recreate from any trigger.

## Backwards compatibility

- Both options default to today's behavior.
- No change to the `SearchClientAdapter` interface (the comparison uses the existing `search`), and no index mapping or schema change.
- The stock check uses `ProductVariantService.getSaleableStockLevel` and the comparison reads through the adapter, so both features behave identically on Elasticsearch and OpenSearch.

## Implementation notes

- `updateProductsOperationsOnly` was split into `buildProductVariantOperations` (build) and a thin execute wrapper, so the incremental path can build the target documents, diff them, and skip the write. Output is unchanged and writes are still chunked via `executeBulkOperationsByChunks`.
- The document comparison lives in `src/indexing/index-diff.ts` (`stableStringify`, `indexedDocumentsMatch`) so it can be unit tested in isolation.

## Tests

Unit (`src/indexing/index-diff.spec.ts`): `stableStringify`, `targetDocumentsById` (pairing the update and doc bulk operations back into a document map), and `indexedDocumentsMatch`, including that a stock-derived custom field difference still counts as a change, which is the correctness invariant behind `onStockStatusChange`.

E2E (`e2e/elasticsearch-plugin.e2e-spec.ts`): the existing suite now runs with `skipUnchangedIndexUpdates` enabled, so every existing test exercises the guard on real updates. A new `skipUnchangedIndexUpdates` describe adds explicit stock-movement cases: a non-boundary movement leaves search correct, and movements that flip `inStock` in and out of stock are reflected. Runs under both backends via `SEARCH_BACKEND`.

I validated `onStockStatusChange` (the pre-enqueue guard) manually against a running OpenSearch instance: a non-boundary movement created no reindex job, a boundary movement flipped `inStock`/`productInStock`, and a full reindex rebuilt everything. I did not enable `onStockStatusChange` in the e2e config because the existing spec defines a stock-derived custom mapping, and that mode is not intended for such setups. Happy to add dedicated coverage for it if you would like, and to adjust the option names or defaults.


packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
```diff
@@ -500,12 +605,17 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
...
     }
 
-    private async updateProductsOperationsOnly(
+    /**
+     * Builds (but does not execute) the upsert operations for a single product's variant
+     * documents. Split out from {@link updateProductsOperationsOnly} so the incremental update
+     * path can build the target documents, compare them against what is currently indexed, and
+     * skip the write entirely when nothing changed (see `skipUnchangedIndexUpdates`).
+     */
+    private async buildProductVariantOperations(
```
biggamesmallworld: Both mid-build flushes are gone, so this buffers every operation for a product before returning. The description says writes are still chunked, which is true, but the *build* isn't anymore.

Count is `channels × languages × variants × 2`, each holding a full `VariantIndexItem`. 5000 variants, 3 channels, 4 languages = 120k buffered docs. `reindex()` at line 388 goes through here, so full reindex on a big catalog can now OOM. No opt-in needed.

The diff path needs them in memory, that's fine, it's opt-in. But `updateProductsOperationsOnly` has to keep streaming. A flush callback would do it.

packages/elasticsearch-plugin/src/indexing/index-diff.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * Deterministic JSON serialization (keys sorted, `undefined` omitted, array order preserved).
+ */
+export function stableStringify(value: any): string {
```
biggamesmallworld: `fast-deep-equal` is already in `package.json` and imported in `elasticsearch.service.ts:19`.

Also: `stableStringify(new Date())` returns `{}`. A custom mapping returning a Date never matches its own `_source` (an ISO string), so the skip never fires and the option quietly does nothing. Tests only cover primitives and plain objects.

`equal(JSON.parse(JSON.stringify(doc)), hit._source)` handles `undefined`, Dates, and key order, and compares exactly what gets stored.

packages/elasticsearch-plugin/src/indexing/index-diff.ts
```diff
@@ -0,0 +1,56 @@
...
+        .filter(k => value[k] !== undefined)
+        .sort();
+    return `{${keys.map(k => `${JSON.stringify(k)}:${stableStringify(value[k])}`).join(',')}}`;
+}
+
+/**
+ * Pairs the `{ update: { _id } }` and following `{ doc }` bulk operations produced for a product
+ * into a map of document id to document.
+ */
+export function targetDocumentsById(
```
biggamesmallworld: `buildProductVariantOperations` has the id and the doc at construction time. Return `{ operations, documentsById }` and this function plus its four tests go away. `Array<{ operation: any }>` also throws away the `index` field.

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
```diff
@@ -193,6 +194,110 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
...
+            return true;
+        }
+    }
+
+    /**
+     * Recomputes a product's per-channel stock booleans and compares them to what is currently
+     * indexed. Returns `true` if any variant `inStock` or the product `productInStock` differs, or
+     * if the product is not indexed yet.
+     */
+    private async productStockStatusDiffersFromIndex(
```
biggamesmallworld: This recomputes stock status differently from `createVariantIndexItem` and they already disagree. `ProductVariant.deletedAt` is a plain `@Column`, not `@DeleteDateColumn`, so line 249 pulls in soft-deleted variants. The builder filters them and also force-disables everything when `product.enabled` is false, which changes `productInStock`.

Both differences fail safe today. The issue is that `isProductIndexUnchanged` gets this right by calling the real builder, and this one will go stale next time `createVariantIndexItem` changes with nothing to catch it.

packages/elasticsearch-plugin/src/indexing/indexer.controller.ts
```diff
@@ -193,6 +194,110 @@ export class ElasticsearchIndexerController implements OnModuleInit, OnModuleDes
...
+        ctx: MutableRequestContext,
+        productId: ID,
+    ): Promise<boolean> {
+        const result = await this.adapter.search({
+            index: this.options.indexPrefix + VARIANT_INDEX_NAME,
+            body: {
+                query: { term: { productId } },
+                _source: ['channelId', 'productVariantId', 'inStock', 'productInStock'],
+                size: 10000,
```
biggamesmallworld: `10000` is the default `index.max_result_window`, so this truncates at the ceiling with no error. Same literal at 795.

Safe there (fewer hits = mismatch = write), not here: dropped hits mean dropped channel buckets, `if (!channelDocs) continue` skips them, guard says unchanged. Named constant plus handle the boundary. Both call sites are the same `term: { productId }` search too.

packages/elasticsearch-plugin/e2e/elasticsearch-plugin.e2e-spec.ts
```diff
@@ -1698,6 +1699,58 @@ describe(`Elasticsearch plugin [${searchBackend as string}]`, () => {
             });
         });
     });
+
+    // The whole suite runs with skipUnchangedIndexUpdates enabled (see the plugin config above),
+    // so every existing test already exercises the guard on real updates. These add explicit
+    // stock-movement cases: a non-boundary movement must leave search correct, and a movement
+    // that flips inStock must be reflected.
+    describe('skipUnchangedIndexUpdates', () => {
```
biggamesmallworld: These pass with the flag off: they assert search is correct, which it already was. Nothing checks a write was skipped.

Sentinel makes it deterministic: write an extra field into the index doc, run the no-op update, assert it survived. Proves the delete-then-recreate didn't happen.

`it('finds an in-stock variant to exercise')` is a `beforeAll` in disguise, and the next two break with `undefined` if it fails. Fixture is fixed, so assert the exact variant. `take: 200` will silently truncate once the fixture grows.

Flipping the whole suite also means these 96 tests stop covering the default path. The uuid spec still does, but that's 7 tests. Prefer a targeted describe.

biggamesmallworld (changes requested):
Nice approach on `skipUnchangedIndexUpdates`. Diffing against the real builder output is the right call.

Three blockers:

1. The refactor dropped bulk-operation chunking. Hits everyone, including full reindex, with both options off. See inline.
2. Please drop `reindexOnStockMovement`. Your description says `skipUnchangedIndexUpdates` is correct for any config and covers any trigger, which makes the other one a subset that's only correct sometimes. Add a stock-derived custom mapping later and you get a silently stale index. It also does an ES search plus a Product query plus per-variant `getSaleableStockLevel` in the subscriber, so order placement pays for it. Dropping it takes `productStockStatusDiffersFromIndex` and half the PR with it.
3. `stableStringify` reimplements `fast-deep-equal`, which we already depend on, and it's buggy. See inline.

Also missing `@since` on both options, and `elasticsearch-options.mdx` needs regenerating.

LeftoversTodayAppAdmin:
Thanks for the thorough review. I reworked the PR around what turned out to be the real issue, the delete-then-recreate, and made every new behaviour opt-in, so with the defaults unchanged the plugin behaves exactly as before. Nothing changes for current installs until a flag is set.

**Two opt-in options.**

- `incrementalIndexUpdates` (default false). When on, an update reads the product's current documents, compares them against the freshly built ones, and writes only the difference: it replaces changed and new documents in place, deletes the ones that no longer exist, and writes nothing when nothing changed. A live document is never removed first, so the product stays searchable during an update. It compares the whole document, so it is correct for any mapping config. A full reindex still writes everything, and a product too large to diff falls back to delete-then-recreate. With the flag off, the write path is the historic one.
- `reindexOnStockMovement` (default 'always'). With 'onStockStatusChange' the plugin checks, before enqueuing, whether a stock change would flip inStock or productInStock and skips the job when it would not. This covers both order-driven StockMovementEvents and admin variant updates that touched only stock fields, since ProductVariantEvent carries the update input.

**Your specific points.**

1. Chunking and reindex OOM. Fixed. The builder is now a visitor that yields one document at a time, and the reindex and default paths stream and flush in chunks, so a full reindex never buffers a whole large product. Only the incremental path collects to diff, capped at the result window with a fallback to streaming.

3. stableStringify. Gone, replaced by the existing fast-deep-equal, comparing the built document as the index stores it. The JSON round-trip handles undefined, Dates and key order, which was the Date bug. The diff is a pure function in index-diff.ts with unit tests. targetDocumentsById is gone too, since the builder hands the id and document straight to the diff.

Magic number. Both index reads use a named INDEX_MAX_RESULT_WINDOW and fall back to a full write if a read reaches the window.

@since and docs. Added to both options. Documented both options and how to enable them on the plugin page and regenerated the reference docs. Heads up that the regen also picked up earlier drift from the adapter refactor, since the committed docs were stale.

Tests. The default suite runs with both flags off, so it still covers the historic path. incrementalIndexUpdates has its own spec that reads the document directly rather than paging with take: 200, asserts the exact fixture variant, and proves a skip by checking the document _version stays put on a no-op and increments on a real change. A separate stock-guard spec with no custom mappings proves a non-flipping admin stock edit creates no write while a flipping or non-stock edit does. I ran the full suite locally against both Elasticsearch 9.3.4 and OpenSearch 3.7.0, all green.

**On point 2, keeping reindexOnStockMovement.** With correctness owned by the whole-document paths, this is a pre-enqueue optimization. It self-disables when a custom mapping is configured, so the stock-booleans-only check runs only when those booleans are provably the only stock-derived fields, and adding a mapping later cannot cause a stale index. The recompute that could drift is gone, both paths share one helper, and it no longer does an index search or a heavy product load on the order path. It is not redundant with the write-path guard: that one runs inside the worker, so by the time it decides nothing changed, the job row is written, the poll waited, and the worker ran. On a database-backed queue that per-change job is the dominant cost, and stock changes are the highest-frequency event in a busy catalog, so skipping the job keeps the queue shallow and lets real updates reach the index sooner.

Thanks again for the careful read. Everything is opt-in, so it is safe to merge and enable per deployment.

LeftoversTodayAppAdmin:
Hi @biggamesmallworld please review the latest changes! Thank you.

